PR #139, opened 2026-04-11
client: drop poison panics and surface invite parse errors

### LINKED ISSUE #114: [client] Fix three lock().unwrap() poison vectors on join_links
Parent: #108

## Problem

Three sites in the client use `std::sync::Mutex::lock().unwrap()` on `Arc<Mutex<Vec<JoinLink>>>`:

- `crates/client/src/joining.rs:206` — `self.join_links.lock().unwrap().push(link);`
- `crates/client/src/joining.rs:211` — `self.join_links.lock().unwrap().clone()`
- `crates/client/src/listeners.rs:333` — `let mut links = ctx.join_links.lock().unwrap();`

If any prior panic happens while any of these hold the lock, the mutex is poisoned. Every subsequent caller then unwraps an `Err` and panics. `listeners.rs:333` is on the hot path for inbound `JoinRequest` messages — one poisoned lock permanently downs all future join-request handling.

## Fix

Two options, pick one:

**Option A (recommended): switch to `parking_lot::Mutex`.** It doesn't poison and `lock()` returns the guard directly. This is the smallest diff and the best ergonomics. `parking_lot` is already a common dep in Rust projects and would not meaningfully increase the dependency graph.

**Option B: handle poison explicitly.** Replace each `.lock().unwrap()` with a helper that recovers from poison and logs:

```rust
fn lock_or_recover<'a, T>(m: &'a Mutex<T>, ctx: &str) -> MutexGuard<'a, T> {
    match m.lock() {
        Ok(g) => g,
        Err(poisoned) => {
            tracing::error!(ctx, "join_links mutex was poisoned; recovering");
            poisoned.into_inner()
        }
    }
}
```

## Test

Add a unit test that simulates panic-while-holding:

```rust
#[test]
fn poisoned_join_links_mutex_does_not_crash_next_caller() {
    let links = Arc::new(Mutex::new(Vec::<JoinLink>::new()));
    let links2 = Arc::clone(&links);
    let _ = std::panic::catch_unwind(|| {
        let _guard = links2.lock().unwrap();
        panic!("simulated panic while holding lock");
    });
    // next caller should still work
    let mut guard = lock_or_recover(&links, "test");
    guard.push(/* a join link */);
}
```

### LINKED ISSUE #115: [client] Propagate UUID parse errors during invite join instead of silently regenerating
Parent: #108

## Problem

`crates/client/src/joining.rs:68-69`:

```rust
server.id = willow_channel::ServerId(
    uuid::Uuid::parse_str(&server_id).unwrap_or_else(|_| uuid::Uuid::new_v4()),
);
```

And `crates/client/src/joining.rs:76`:

```rust
let ch_id = server
    .create_channel(name, willow_channel::ChannelKind::Text)
    .unwrap_or_else(|_| willow_channel::ChannelId::new());
```

If the invite carries a non-parseable `server_id` or `create_channel` fails, the joiner **silently generates a fresh UUID and joins as if the invite were valid**. The joiner now thinks they're in a different server from everyone else — a near-impossible-to-diagnose split-brain.

## Fix

1. Define (or reuse) an error variant `ClientError::MalformedInvite(String)` with a descriptive reason.
2. Replace `.unwrap_or_else(|_| Uuid::new_v4())` with `?` propagation that returns `ClientError::MalformedInvite(format!("invalid server_id: {e}"))`.
3. Do the same for the `create_channel` call — failing to create a channel during invite acceptance should surface to the user, not silently create a new channel ID that no one else knows about.
4. Grep `crates/client/src/` for other `unwrap_or_else(|_| Uuid::new_v4())` or `unwrap_or_default()` patterns on parsed IDs and fix them too.

## Test

Add a test to `crates/client/src/lib.rs` test module:

```rust
#[tokio::test]
async fn malformed_invite_server_id_is_rejected() {
    let client = test_client();
    let bad_invite = InviteToken { server_id: "not-a-uuid".into(), /* ... */ };
    let result = client.accept_invite(bad_invite).await;
    assert!(matches!(result, Err(ClientError::MalformedInvite(_))));
}
```

intendednull:
## Summary

Two small robustness fixes in `willow-client`, both pulled from the review under #108:

- **#114** — drop `std::sync::Mutex::lock().unwrap()` on `join_links` so a panic while holding the guard no longer poisons the lock and takes down every future caller (including the gossip listener that handles `JoinRequest`). The `Arc<Mutex<Vec<JoinLink>>>` now uses `parking_lot::Mutex`, which is poison-free by construction. `parking_lot` is added once at the workspace level and pulled into `crates/client` only.
- **#115** — `accept_invite` previously used `unwrap_or_else(|_| Uuid::new_v4())` / `unwrap_or_else(|_| ChannelId::new())` to silently invent fresh IDs when an invite's `server_id` or a channel name failed to parse. That split-brained the joiner from every other peer. Introduce `ClientError::MalformedInvite(String)` (a new `thiserror` variant — the first typed error in the crate) and propagate parse / create-channel failures with `?`, so the caller sees a clear `anyhow::Error` that downcasts to `ClientError::MalformedInvite` instead of a phantom server membership.

Scope is strictly `crates/client/` plus the two `Cargo.toml` entries for `parking_lot` and `thiserror`. No other crates touched.

Closes #114
Closes #115
Progresses #108

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace -- -D warnings` — clean, zero warnings
- [x] `cargo test -p willow-client` — 68 tests pass including the two new regression tests:
  - `join_links_lock_survives_panic_in_holder` — spawns a `spawn_blocking` task that pushes a link, panics while holding the guard, then asserts the main task can still read/write `join_links` without panicking.
  - `malformed_invite_server_id_is_rejected` — builds a real invite encrypted for the test client, tampers the embedded `server_id` to `"not-a-uuid"`, re-encodes, and asserts `accept_invite` returns `ClientError::MalformedInvite` mentioning `server_id`.
- [x] `cargo test -p willow-state -p willow-channel -p willow-network` — all upstream dependents still green.
- [x] `cargo check --target wasm32-unknown-unknown` for the full `just check-wasm` cohort (`-p willow-identity -p willow-state -p willow-channel -p willow-messaging -p willow-crypto -p willow-transport -p willow-common -p willow-network -p willow-client -p willow-web`) — clean, verifying `parking_lot` builds on WASM.

https://claude.ai/code/session_018TTGhL645aTR4RZWrRBLPS


intendednull:
Audit note: this PR and #138 both edit `crates/client/src/joining.rs` in the same ~10-line region (the `else` branch that constructs a `Server` from an accepted invite). The overlap is real — whichever lands second needs a manual rebase.

Planned merge order: **#138 first, then rebase this PR on top of main.** The rebase will replace the `let mut server = Server::new(...); server.id = ServerId(parsed_server_uuid);` pair with a single `Server::with_id(ServerId(parsed_server_uuid), &accepted.server_name, accepted.genesis_author)` call — the upfront UUID validation and the `create_channel` `map_err` from this PR both stay.

Also noting for a follow-up (out of scope here): `crates/client/src/lib.rs` still has a `Uuid::parse_str(...).unwrap_or_else(|_| Uuid::new_v4())` pattern in the channel-id reconciliation path. Different context (reading stored IDs rather than untrusted invite payloads), but the same split-brain shape as #115. Tracked in a follow-up issue.

---

PR #221, opened 2026-04-24
spec: outbox-style per-peer relay discovery

intendednull:

## What & why

Willow currently has no peer-discovery mechanism beyond hardcoded bootstrap relay addresses. Share links carry a single relay URL, and clients have no way to find other workers that serve the same server.

Nostr's NIP-65 "outbox model" solves this without a DHT: each user publishes a `kind:10002` replaceable event listing their relays. To fetch **from** a user, read their write relays; to reach them, publish to their read relays. When publishing, push the user's 10002 along so discovery spreads.

This spec proposes a Willow equivalent — new `EventKind::RelayList` carrying signed `RelayEntry` records:

```rust
struct RelayEntry {
    url: String,                  // ws://, tcp://, or iroh NodeId
    role: RelayRole,              // Read | Write | ReadWrite
    capability: RelayCapability,  // Relay | ReplayWorker | StorageWorker
    weight: Option<u8>,
}
```

Published on `_willow_profiles` (public, per-peer-global) and optionally inside the server DAG (members-only, per-server). Replaceable semantics use per-author `seq` monotonicity — side-stepping NIP-65's clock-skew LWW problem entirely.

**Authority stays with `SyncProvider` grants** — the outbox is strictly a routing hint, not a trust signal. Clients prefer entries whose `EndpointId` also holds a live grant. This inversion is exactly what Nostr keeps reinventing (NIP-29's "relay is authority" failure mode).

Spec covers share-link bootstrap flow, anti-centralization client heuristics (latency-weighted, operator-diversified rotation), privacy via encrypted per-server lists, and tests across state / client / e2e layers.

Spec file: [`docs/specs/2026-04-24-outbox-relay-discovery.md`](https://github.com/intendednull/willow/blob/claude/spec-outbox-relay-discovery/docs/specs/2026-04-24-outbox-relay-discovery.md)

## Open questions for review

1. Scope — per-peer-global, per-server, or both?
2. Public vs encrypted RelayList — doxxing tradeoff vs discovery usability
3. Liveness: NIP-66-style bot-published liveness events, or rely on client probing?
4. How do we actively prevent the de-facto centralization Nostr suffers from (damus.io, nos.lol, etc.)?
5. Should `SyncProvider` grant holders be *required* to publish a RelayList?
6. Should `willow-state` introduce a general "replaceable event kinds" mechanism, or is this one-off sufficient?

- Bech32 identifiers (separate PR): `wrelay` HRP fits here
- Relay capability doc (separate PR): clients use RelayList to find candidates, then fetch capability doc to pick
- Epoch rotation: encrypted RelayList uses current epoch key
- Existing `docs/specs/2026-03-27-shareable-join-links-design.md` — interaction with share-link bootstrap

---

---

### DIFF
--- a/docs/specs/2026-04-24-outbox-relay-discovery.md
+++ b/docs/specs/2026-04-24-outbox-relay-discovery.md
@@ -1,0 +1,253 @@
+# Outbox-style Relay Discovery
+
+> **One-sentence summary:** each peer publishes a signed, replaceable
+> `RelayList` event advertising which relays and workers carry its
+> data, so clients can discover the rest of the network from a single
+> bootstrap URL — no global DHT, no hardcoded relay fleet, and no trust
+> promotion beyond the existing `SyncProvider` grant.
+
+## Motivation
+
+Willow's share links today carry a single bootstrap relay URL (see
+`docs/specs/2026-03-27-shareable-join-links-design.md`) and
+`IrohNetwork` otherwise relies on a hardcoded relay list baked into
+`crates/network/src/iroh.rs`. This has three failure modes:
+
+1. **Centralization.** If the hardcoded relay is down, new joiners
+   cannot complete the `JoinRequest`/`JoinResponse` handshake.
+2. **No graceful migration.** An owner who moves their worker fleet
+   (e.g. Linode → Fly.io) must reissue every share link.
+3. **No worker discovery.** A server can have a `SyncProvider`-granted
+   replay or storage worker, but a joining client has no way to learn
+   its network address short of the owner encoding it manually.
+
+Nostr solves an analogous problem with NIP-65 "outbox model" relay
+lists: each user publishes a `kind:10002` replaceable event listing
+their read/write relays. Willow already has the substrate to do
+this — signed events on `_willow_profiles`, an HLC for ordering,
+and a cryptographic trust chain rooted at the server owner.
+
+## Relationship to `SyncProvider`
+
+The outbox is a **routing hint, not a capability grant.** Authority
+still flows exclusively through `EventKind::GrantPermission { permission: SyncProvider }`
+as defined in `crates/state/src/event.rs:22` and enforced by
+`apply_event` in `crates/state/src/materialize.rs:130`. A `RelayList`
+advertising `wss://evil.example` as a replay worker does not make
+that host trusted — it's only a *candidate address* that clients
+attempt to connect to. Once connected, the usual authority check
+applies: the peer at that address must sign state updates with a key
+that has a live `SyncProvider` grant in the server DAG.
+
+| Layer | Who decides | Enforced where |
+|-------|-------------|----------------|
+| "Which addresses might serve this server?" | Any peer's `RelayList` | Client connection code (advisory) |
+| "Is this peer allowed to sync state?" | Owner / admin via `GrantPermission` | `required_permission()` + `apply_event` |
+| "Should I prefer this address?" | Client heuristics (weight, latency, trust intersection) | `willow-client` discovery loop |
+
+Clients **SHOULD** prefer outbox entries whose advertised `EndpointId`
+is also the target of a current `SyncProvider` grant in the server's
+DAG. This is what turns the outbox from a bare hint into a
+trust-weighted routing table.
+
+## New event kind
+
+Add a variant to `EventKind` (`crates/state/src/event.rs:70`):
+
+```rust
+EventKind::RelayList {
+    entries: Vec<RelayEntry>,
+    replaces: Option<EventHash>, // previous RelayList by same author
+}
+
+pub struct RelayEntry {
+    /// Wire address. Accepts `ws://`, `wss://`, `tcp://`, `quic://`,
+    /// or a bare iroh `NodeId` ("iroh:<base32-nodeid>").
+    pub url: String,
+    /// What direction the relay is useful for (NIP-65 analog).
+    pub role: RelayRole,
+    /// What the remote peer actually runs.
+    pub capability: RelayCapability,
+    /// Client-side ranking hint. 0 = lowest, 255 = highest. None = default.
+    pub weight: Option<u8>,
+}
+
+pub enum RelayRole { Read, Write, ReadWrite }
+pub enum RelayCapability { Relay, ReplayWorker, StorageWorker }
+```
+
+Authority tier: **unrestricted.** Any member may publish their own
+`RelayList`. Trust comes from the signature + existing
+`SyncProvider` grants, not from membership in a new permission tier.
+Add to the intentionally-unrestricted comment block in
+`required_permission()` (`crates/state/src/materialize.rs:283`).
+
+## Replaceable semantics
+
+NIP-65 treats `kind:10002` as a replaceable event: the newest
+`created_at` from a given author supersedes older ones. Willow's
+per-author Merkle DAG already gives us something stronger — a
+monotonic `seq` (`crates/state/src/event.rs:191`) plus HLC-ordered
+`timestamp_hint_ms`. The rule is:
+
+> For any `(author, EventKind::RelayList)`, the event with the
+> **highest `seq`** in that author's chain is authoritative. Older
+> `RelayList` events are retained in the DAG for audit but are
+> ignored by `apply_mutation`.
+
+This is the first "replaceable kind" in `willow-state`. **Flag for
+reviewers:** we either (a) introduce a generic `ReplaceableKind` trait
+keyed on `(author, discriminant)` so future additions (e.g.
+`SetProfile` arguably belongs here) share the machinery, or (b) just
+add a `state.relay_lists: HashMap<EndpointId, RelayList>` and special-
+case it. Option (a) is cleaner but a wider change; option (b) is the
+minimum viable path and is what this spec recommends shipping first.
+
+Because `seq` is totally ordered within an author chain, replaceable
+semantics in Willow are **not** vulnerable to the clock-skew
+last-write-wins pathology that NIP-65 inherits.
+
+## Scope: per-peer-global vs per-server
+
+| Axis | Per-peer-global (`_willow_profiles`) | Per-server (inside server DAG) |
+|------|--------------------------------------|--------------------------------|
+| Discoverable from a share link alone | Yes — single lookup by owner PeerId | No — client must already have genesis event |
+| Doxxes your worker fleet to the world | Yes | No — visible only to members |
+| Follows key rotations | Yes, tied to peer identity | Yes, but requires DAG access |
+| Natural fit for `SyncProvider` nominations | No — grants are per-server | Yes |
+| Bootstrap-before-membership usable | Yes | No |
+
+**Recommendation:** publish two `RelayList` streams.
+
+1. A **public per-peer-global** `RelayList` on `_willow_profiles`,
+   used by joiners who only have a share link and need to find the
+   owner. Entries here are treated as hints about the *peer*, not
+   about any specific server.
+2. A **per-server** `RelayList` inside the server DAG (so it merges
+   with the rest of server state), used to nominate the server's
+   `SyncProvider` workers. Members read this after joining.
+
+The per-server stream MAY be published encrypted (see Privacy
+below); the per-peer-global stream is inherently public because its
+job is to bootstrap strangers.
+
+## Discovery flow
+
+New joiner clicks a share link (`JoinToken` per
+`2026-03-27-shareable-join-links-design.md`):
+
+1. Connect to the bootstrap relay embedded in the link.
+2. Subscribe to `_willow_profiles` (`crates/network/src/topics.rs:29`)
+   and request the owner's latest `RelayList` event by
+   `(author = inviter_peer_id, kind = RelayList)`.
+3. In parallel, request the server DAG from the bootstrap relay and
+   replay to obtain the set of current `SyncProvider` grants.
+4. **Intersect:** compute the set of `RelayEntry` whose advertised
+   `EndpointId` also holds a live `SyncProvider` grant. These are
+   trust-weighted candidates.
+5. Rank candidates by `(is_trusted, weight, observed_latency)` and
+   connect to the top N (default N=3, following NIP-65's "2–4 each"
+   guidance).
+6. Subscribe to `_willow_profiles` with a long-lived watch so that
+   new `RelayList` events from the owner (key rotation, fleet
+   migration) update the candidate set without a restart.
+7. When a peer publishes a new message, `willow-network` MUST push to
+   its *own* write relays and also opportunistically to the
+   recipient's read relays — mirroring NIP-65's "publish to both
+   sides" rule so discovery spreads without a central indexer.
+
+## Liveness
+
+Two options, analogous to NIP-66:
+
+- **Client-side probing (ship first).** Clients track connect
+  success/failure per entry in memory, decay weights on failure, and
+  write nothing back to the network. Simple, private, fully
+  distributed. Downside: every joiner re-learns the hard way.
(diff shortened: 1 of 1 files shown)


intendednull (commented):
Thanks for a spec that is refreshingly clear about what it is *not* doing. The "routing hint, not capability grant" framing (lines 32-41) is the most important paragraph in the whole research set — it neutralises the usual "but can't an attacker advertise evil workers?" objection by punting trust entirely to `SyncProvider`. Good.

That said, I have structural concerns. The biggest is the "replaceable kind" concept — it's introduced here as a minor implementation note but it's actually a new *category* of event semantics in `willow-state`, and I don't think the spec grapples with the consequences. Details below.

## Strengths

- **Authority separation is crisp.** The table at lines 42-46 makes the hint-vs-grant distinction concrete and maps each layer to a specific enforcement site in the codebase. This is the right design.
- **Uses existing substrate.** `_willow_profiles` + per-author `seq` already exist; no new topic, no new crypto, no new trust tier. Good leverage.
- **`seq`-based replacement beats NIP-65.** Lines 86-108 correctly identify that Willow's monotonic per-author `seq` is a strictly stronger ordering than NIP-65's `created_at`; the clock-skew last-write-wins pathology genuinely does not apply here. This deserves to be flagged as a real advantage over the Nostr analog.
- **"Intersect outbox with live `SyncProvider` grants"** (lines 48-51) is the single most important client-side rule in the spec — it's what turns a hint into a trust-weighted routing table. Keep this prominent.
- **Privacy section is honest** about the doxxing tradeoff NIP-65 failed on, and the two-scope (public-minimal + per-server-encrypted) split is a real mitigation.

## Concerns

### 1. "Replaceable kind" is a bigger architectural decision than the spec presents

The spec frames this as a choose-your-own-adventure between a `ReplaceableKind` trait (option a) and a one-off `relay_lists: HashMap<...>` map (option b), and recommends (b). But `apply_event` in `crates/state/src/materialize.rs:130` is a **pure function over the full event stream** — every event in the DAG is re-applied on every `materialize()`. Today there is no concept of an event being "superseded and silently dropped"; `SetProfile` simply overwrites on each replay (lines 471-480 of `materialize.rs`), which is *idempotent replacement*, not *supersession*.

Saying "older `RelayList` events are retained in the DAG for audit but ignored by `apply_mutation`" (line 95) is easy to type but surprisingly load-bearing. Concrete questions:

- **During replay**, do you apply all `RelayList`s and let the last-by-seq win (cheap but wasteful)? Or do you scan once to find the max-seq `RelayList` per author, then only apply that one during the replay loop (requires a two-pass replay — a real departure from the current single-pass design)?
- **Does this interact with `StateHash`?** If two peers have different knowledge of a superseded `RelayList` but agree on the latest, do they converge to the same `StateHash`? If `relay_lists` is part of `ServerState` and the older events exist in one replica's event set but not the other's, the replay will visit different event sequences but (if implemented correctly) produce the same map — OK. But this needs to be stated explicitly, with a convergence test.
- **Merge determinism (`merge()`).** With per-author `seq` strictly monotonic, a single author can't have two `RelayList`s at the same `seq`. But two *devices* of the same author (see concern 4 below) can — and the spec doesn't say which wins.

I'd rather see this spec either (i) commit to option (a) now and design the generic mechanism, documenting `SetProfile` as a future retrofit candidate, or (ii) commit hard to option (b) and explicitly scope "replaceable" to *just this enum variant* with no pretense of generality. Leaving it as "we'll probably generalise later" invites the worst of both worlds.

### 2. Multi-device writes will reject, not merge

Open question #9 from the review brief (not in the spec but worth addressing): if two devices share the same identity and both publish `RelayList` at `seq=N`, the second one is rejected as a duplicate/forked chain — `prev` won't match. The outbox model therefore assumes one writer per author identity. That's fine, but **this should be explicit**: multi-device identity sharing is out of scope, and operators should use distinct `EndpointId`s per device and a single nominated writer (or a future multi-writer design). Otherwise an admin with Linux desktop + phone will silently lose updates.

### 3. Encryption scope for per-server `RelayList` is ambiguous vs PR #220

The spec says "encrypted to the current channel-key membership set (`willow-crypto::seal_content`)" (lines 181-184). But #220 (epoch key rotation) defines `RotateChannelKey` triggered by membership changes, yielding per-epoch keys. Which epoch's key encrypts a per-server `RelayList`?

- If it's "the epoch in force at publication time", a new member can't decrypt a `RelayList` published before they joined — but that's the *whole point* of the outbox, so they need the current one republished every rotation.
- If it's "whatever epoch the reader currently holds", that's a key-agility contract the current `seal_content` doesn't have.

Per #220's "default policy grants new members the current epoch key only", new members literally cannot read stale `RelayList`s. The spec should either (a) mandate that a new `RelayList` is republished at every epoch rotation (bounding per-server outbox lag to one membership event), or (b) reference a dedicated "long-term server-membership key" separate from message-channel epoch keys. Right now the composition is underspecified.

### 4. Anti-centralization section is aspirational, not concrete

The four bullets at lines 206-216 ("never greedy-pick the most common", "weight by latency", "diversify by operator", "rotate periodically") are directionally right but too vague to implement and too vague to test. NIP-65 failed at exactly this level of abstraction.

- "Never greedy-pick the most common entry. If 80% of observed `RelayList`s cite the same URL, reduce its weight on the margin" — reduce by how much? Measured over what window? Does this produce a stable equilibrium or oscillation?
- "Diversify by operator" — Willow has **no operator identity primitive**. `EndpointId` is per-node, but a single operator can run N nodes, and we have no way to tell. Without a way to cluster nodes by operator, "operator diversity" is unimplementable.

Either drop the section, or commit to a concrete algorithm (with pseudocode) and test targets ("in simulation with 1 000 outbox entries and 3 big operators, no single entry's traffic share exceeds 40%"). As written this is hand-waving that will let the exact same damus.io centralization reappear.

### 5. Bootstrap fallback is defeated by the share-link spec

The spec says (lines 195-201) "share links MAY carry a small list (2–3) of bootstrap relays so a single down relay does not block joining — a one-line extension of `JoinToken`." I checked `docs/specs/2026-03-27-shareable-join-links-design.md` — `JoinToken` currently holds `inviter_peer_id`, `server_id`, `link_id`, `server_name`, `inviter_name` (lines 33-40). There is **no bootstrap relay URL field at all**. The shareable-join-links spec implicitly relies on the hardcoded relay baked into `IrohNetwork`.

So "MAY carry a list of bootstrap relays" is actually "MUST add a bootstrap field to `JoinToken` first". That's a cross-spec dependency, not an optional extension. This should either be scoped into this PR explicitly ("amend `JoinToken` to add `bootstrap_relays: Vec<String>`, default empty") or filed as a separate prerequisite issue. Leaving it as "MAY" understates the work.

### 6. Liveness recommendation is defensible but should take a firmer stance

The review brief asks for a position. Mine: **liveness is the client's problem** (try and retry with decay), not the protocol's. `EventKind::RelayLiveness` is a classic Nostr protocol-over-reach — it encodes temporary network state into the permanent signed DAG, which is a category error. The existing HLC bound plus client-side decay give you everything you need; `RelayLiveness` adds DAG bloat and a coordination problem (who's the authoritative prober?) without solving anything. The spec recommends client probing first and I endorse that — but delete the "if client-side proves insufficient" escape hatch. Commit to it.

### 7. Mandatory publication (open question #5): voluntary

`SyncProvider` grant holders should *not* be required to publish a `RelayList`. Mandatory publication creates a split-brain enforcement problem (where? the state machine can't enforce "you must also emit event X") and pressures operators to expose endpoint info when they might legitimately want to run in a private/mesh configuration. Voluntary + client-side fallback to hardcoded bootstrap + share-link bootstrap is strictly better.

## Suggestions

- **Pick option (a) or (b) for replaceable-kinds now, and write the state-machine code sketch for whichever you choose** (not just the data model). Specifically, show what `apply_mutation` looks like when it encounters a superseded `RelayList`, and write a convergence test covering "replay set A+B vs replay set B alone produces the same `ServerState.relay_lists`".
- **Add a concrete multi-device policy** — even if it's just "one writer per identity; divergent devices get rejected at merge; future multi-device is a separate spec".
- **Coordinate with #220 on encryption.** Add a section explaining which epoch key encrypts per-server `RelayList`, or explicitly defer per-server encryption to a follow-up until #220 lands.
- **Make anti-centralization testable** — or drop it. Concrete algorithm + simulation target, or delete the aspirational bullets.
- **Scope the `JoinToken` extension into this PR** (or make it a hard prerequisite). Don't leave bootstrap fallback as "MAY".
- **Resolve open questions.** Six open questions at the bottom is a lot for a spec that's otherwise this well-reasoned. Suggest closing 3, 4, 5 outright (client probing, demote-not-refuse, voluntary) and converting 1, 2, 6 into explicit staged-rollout commitments.
- **Test coverage gaps.** The tests table doesn't cover: (i) convergence across peers with different subsets of superseded `RelayList`s, (ii) `seq` monotonicity violation (forked chain rejection), (iii) per-server encrypted `RelayList` decryption across an epoch rotation. Add these.

Overall: the core idea is right and the authority-vs-hint separation is the cleanest part of the Nostr-inspired research set so far. But the replaceable-event semantics and the cross-spec couplings (bootstrap tokens, epoch keys) need to be made concrete before this is implementable.

---

intendednull (commented):
## Round 2 review: comparative prior art (non-Nostr)

Round 1 covered NIP-65 inheritance issues. This pass widens the lens: we are not the first system to need "given a peer identity, find a server it speaks to," and the prior art is uniformly more conservative than what this spec proposes.

### Matrix: `/.well-known/matrix/server` + SRV is the production design NIP-65 imitates

The Matrix federation discovery cascade — `https://<host>/.well-known/matrix/server` (RFC 8615), then `_matrix-fed._tcp.<host>` SRV, then A/AAAA — is the most-deployed federated discovery system on the internet. Two things it gets right that NIP-65 (and this spec) do not:

1. **The bootstrap hint and the rich server-info live in different places.** SRV/well-known is a tiny redirect. The capability/version/auth document is a separate fetch. This composes naturally with **PR #215's capability doc** — that PR is essentially the Willow analog of `/.well-known/matrix/server`. We should not invent a parallel `RelayList` event when #215 already gives us the right shape.
2. **DNS+TLS is the trust anchor, not a signed event.** Matrix relies on existing PKI rather than minting a new replaceable-event semantic. Willow gets this for free via TLS on relays plus iroh's signed NodeAddrs.

If we want a Matrix-shaped design the spec should say so explicitly: "share link carries SRV-or-well-known equivalent; PR #215 capability doc carries the rest."

### did:plc: signed + updateable, with key rotation as a first-class concept

did:plc resolution chains genesis → update operations, each signed by reconfigurable rotation keys, with a content-addressed history. Willow's per-author `seq` chain already gives us this (and the spec correctly notes it sidesteps NIP-65's clock-skew LWW). Worth calling out the **failure mode** that did:plc actually hit in production: the "central directory server" (`plc.directory`) became a single point of trust because clients lazily query it instead of validating the signed chain. The Willow analog is: clients SHOULD verify the `RelayList` is the highest-`seq` event from that author, but if they only query one relay they will believe whatever that relay returns. The spec needs an explicit "verify against ≥N relays" rule, mirroring what the AT Proto spec ended up adding.

### Scuttlebutt pubs: the closest prior art, and the most cautionary

SSB pubs were exactly "per-user list of servers that carry my data," and the project explicitly retreated from them because:
- Pubs that accept invites get overloaded with strangers' feed data.
- Centralization happened anyway — users picked the most-available pub.
- The community moved to **SSB Rooms**, which deliberately *do not* host feeds: they are connection-brokering only, with open invite codes for stranger onboarding.

Willow's relays are already closer to Rooms than to pubs (the relay bridges peers, the workers do the storage). The spec should lean into that distinction: the `RelayList` should advertise **rooms-style brokers** for bootstrap and keep storage workers as a private, post-membership concern. The current spec already half-says this with the public/per-server split — this comparison is reinforcement.

### libp2p rendezvous / iroh pkarr: the "do nothing new" option

This is the most important finding of Round 2. **iroh already has global node discovery** via pkarr (public-key-addressable DNS records over the BitTorrent mainline DHT) — see [Iroh global node discovery](https://www.iroh.computer/blog/iroh-global-node-discovery) and [iroh-pkarr-node-discovery](https://docs.rs/iroh-pkarr-node-discovery/latest/iroh_pkarr_node_discovery/). Each iroh node publishes a signed DNS packet containing its NodeId, relay URL, and direct addresses. Discovery works by NodeId alone.

This means the entire `RelayList` event may be solving a problem we already have a primitive for:

- "Find the owner from a share link" → resolve owner's NodeId via pkarr, get NodeAddr, connect.
- "Owner migrated relay providers" → owner re-publishes pkarr packet, clients resolve fresh.
- "Worker discovery" → workers publish their own pkarr packets; the per-server DAG only needs to list `EndpointId`s (already does via `SyncProvider` grants).

The argument for `RelayList` over pkarr is **role/capability advertisement** (Read/Write/ReadWrite, ReplayWorker vs StorageWorker) — but that is exactly what PR #215's capability doc carries. So the layering becomes:

| Layer | Mechanism | Source of truth |
|---|---|---|
| NodeId → addresses | iroh pkarr | DHT, signed by node |
| EndpointId → role/capabilities | PR #215 capability doc | served by the endpoint itself |
| EndpointId → trusted? | `SyncProvider` grant in server DAG | already exists |

There's no remaining job for `RelayList`. **Strongly recommend the spec be reworked to compose pkarr + #215 instead of inventing a new replaceable event kind.**

### Anti-centralization in practice: honest report

The spec lists protocol-layer mitigations against relay centralization. Empirically, none of them have worked at meaningful scale anywhere:

- **Matrix**: matrix.org carries a dominant share of federation traffic despite ~10K federateable servers existing ([This Week in Matrix 2025-10-03](https://matrix.org/blog/2025/10/03/this-week-in-matrix-2025-10-03/) reports 28.6% publish room directories).
- **ActivityPub**: mastodon.social centralization is a years-running discussion.
- **Nostr**: damus.relay and a handful of others dominate; the [Nostrify outbox docs](https://nostrify.dev/relay/outbox) acknowledge "everyone picking the same few relays" as the problem the model was meant to solve, and benchmarks show it didn't.
- **SSB**: pubs centralized; rooms-with-no-feed-storage is the current answer.

The pattern is clear: **users pick the most-available server regardless of protocol affordances.** The spec should drop the "anti-centralization" framing for `RelayList` — it cannot deliver that — and reframe purely as "graceful migration + worker discovery." Honest scope is more defensible than aspirational scope.

### Recommendation

Rewrite, in priority order:

1. **Stop inventing.** Remove `EventKind::RelayList`. Use iroh pkarr for NodeId→addresses and PR #215's capability doc for what each endpoint runs.
2. If a per-server "trusted workers" hint is still wanted, put it inside the server DAG as a **list of EndpointIds** (no URLs, no roles, no weights) — clients resolve EndpointIds via pkarr.
3. Adopt Matrix's bootstrap shape for share links: a tiny well-known-style hint, with the rich data fetched from the endpoint itself.
4. Drop the anti-centralization claims; reframe as migration + discovery.
5. If we ignore (1) and ship `RelayList` anyway, mandate ≥N-relay verification for replaceable-event reads (lesson from did:plc / `plc.directory`).

Round 1 said "the plan as written is salvageable but needs work." Round 2 is more pointed: **the right move is probably not to ship this event at all** and to reuse infrastructure we already depend on.

---

*Sources: [Matrix Server-Server API](https://spec.matrix.org/latest/server-server-api/) - [did:plc spec](https://github.com/did-method-plc/did-method-plc) - [SSB Rooms announcement](https://www.manyver.se/blog/announcing-ssb-rooms/) - [Iroh global node discovery](https://www.iroh.computer/blog/iroh-global-node-discovery) - [Iroh DNS dial-by-NodeID](https://www.iroh.computer/blog/iroh-dns) - [Nostrify outbox model](https://nostrify.dev/relay/outbox) - [libp2p rendezvous](https://docs.libp2p.io/concepts/discovery-routing/rendezvous/)*

---

PR #280, opened 2026-04-24
fix(client): enforce CreateInvite check, use signer for JoinRequest, correct genesis_author

### LINKED ISSUE #225: [SEC-A-02] `generate_invite` skips the `CreateInvite` permission check
Commit: `2f26d91` · Finding: `SEC-A-02`

### Problem
`ClientHandle::generate_invite` at `crates/client/src/joining.rs:4-72` unilaterally emits a `GrantPermission { permission: SendMessages }` for the recipient without verifying the caller holds `CreateInvite` (or is an admin). The sibling `create_join_link` at `crates/client/src/joining.rs:213-225` DOES perform that check:

```rust
es.has_permission(&pid, &willow_state::Permission::CreateInvite)
```

A non-admin member calling `generate_invite` produces an invite that ships a `GrantPermission` event other peers will reject (because `check_permission` for `GrantPermission` requires admin at `crates/state/src/materialize.rs:103-112`), but the invite artifact itself is propagated. Result: confusing UX at best, permission bypass via a different code path at worst.

### Fix
Mirror `create_join_link`: check `es.has_permission(&pid, &Permission::CreateInvite)` (or `is_admin`) at the top of `generate_invite` and bail with a typed error if false. Add a unit test asserting a non-admin call returns `Err`.

### Obvious fix — will be auto-PR'd.

### LINKED ISSUE #239: [SEC-A-03] `JoinRequest` grants permission to attacker-supplied `peer_id`, not verified signer
Commit: `2f26d91` · Finding: `SEC-A-03` (JoinRequest portion; voice portion is tracked in #172)

### Problem
`crates/client/src/listeners.rs:378-395` handles `JoinRequest { link_id, peer_id }` by using the attacker-supplied `peer_id` field as the grant target rather than the Ed25519 `signer` returned by `unpack_wire`.

Any peer who discovers a valid `link_id` can cause the inviter to hand `SendMessages` (or whatever `GrantPermission` payload the invite carries) to an arbitrary third-party `EndpointId`. Contrast with the chat event path at `listeners.rs:240-242` which enforces `event.author == signer`.

### Fix
Either require `peer_id == signer` (reject mismatch) or drop the `peer_id` field from `JoinRequest` entirely and use `signer` as the grant target.

### Obvious fix — will be auto-PR'd.

### LINKED ISSUE #241: [SEC-A-08] Invite generation picks "genesis author" as `admins.iter().next()` instead of `state.genesis_author`
Commit: `2f26d91` · Finding: `SEC-A-08`

### Problem
`crates/client/src/joining.rs:38-43` computes the "genesis_author" used in the invite as `es.admins.iter().next().copied().unwrap_or(my_id)`. `admins` is a `BTreeSet<EndpointId>` so `.iter().next()` returns the lexicographically smallest EndpointId — which may not be the real genesis author. `listeners.rs:423` consumes this value.

The joiner uses `genesis_author` to identify the owner / root of trust; in a server where the owner's EndpointId happens not to sort first (after later admin grants), the invite will advertise a different peer as the trust root. Since invites are signed hints (not security boundaries), this does not break the security model, but it misleads the joiner's initial sync-provider selection toward a non-owner admin.

### Fix
Use `state.genesis_author` (already authoritative on `ServerState`) rather than `admins.iter().next()`.

### Obvious fix — will be auto-PR'd.

intendednull:
## Summary

Three related client-side auth/signer findings from the recent audit, fixed together in `crates/client/`:

- Closes #225 (SEC-A-02) — `generate_invite` now rejects callers who lack `CreateInvite`, matching the existing guard on `create_join_link`.
- Closes #239 (SEC-A-03, JoinRequest portion) — the `JoinRequest` listener now drops packets whose wire-supplied `peer_id` disagrees with the verified Ed25519 `signer`, and always uses `signer` as the grant target / JoinResponse recipient. Attackers can no longer redirect the invite + `SendMessages` grant at an arbitrary identity.
- Closes #241 (SEC-A-08) — both `generate_invite` and the `JoinRequest` handler now read `state.genesis_author` instead of `admins.iter().next()`, so the invite carries the authoritative owner rather than a non-deterministic admin pick.

## Changes

- `crates/client/src/joining.rs` — `generate_invite`: added `CreateInvite` check up front; switched `genesis_author` to `es.genesis_author.unwrap_or(my_id)`.
- `crates/client/src/listeners.rs` — `JoinRequest` handler: drop on `peer_id != signer`; use `signer` as the grant target and as `JoinResponse.target_peer`; same `genesis_author` correction as above.

## Tests

Four new inline unit tests (all green locally):

- `joining::tests::generate_invite_requires_create_invite_permission` — non-admin caller gets an error naming `CreateInvite`.
- `joining::tests::generate_invite_uses_state_genesis_author` — sets up state where `admins.iter().next() != genesis_author`, then decodes the invite payload and asserts the real owner is carried.
- `listeners::tests::join_request_with_mismatched_peer_id_is_rejected` — Mallory-signed packet naming Bob yields no grant for either identity and does not consume the join link.
- `listeners::tests::join_request_with_matching_signer_grants_send_messages` — positive path still works.

## Test plan

- [x] `cargo check -p willow-client`
- [x] `cargo test -p willow-client` (139 passed)
- [x] `cargo clippy -p willow-client --all-targets -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo check -p willow-client --target wasm32-unknown-unknown`

https://claude.ai/code/session_01N2qhqzCpeTdESRbEsEd71R

---
_Generated by [Claude Code](https://claude.ai/code/session_01N2qhqzCpeTdESRbEsEd71R)_

### DIFF
--- a/crates/client/src/joining.rs
+++ b/crates/client/src/joining.rs
@@ -5,6 +5,18 @@ impl<N: willow_network::Network> ClientHandle<N> {
         &self,
         recipient_peer_id: &willow_identity::EndpointId,
     ) -> anyhow::Result<String> {
+        // Enforce the CreateInvite permission up front so non-authorised
+        // peers cannot mint invites, matching the guard used by
+        // `create_join_link` (see issue #225 / SEC-A-02).
+        let pid = self.identity.endpoint_id();
+        let has_perm = willow_actor::state::select(&self.event_state_addr, move |es| {
+            es.has_permission(&pid, &willow_state::Permission::CreateInvite)
+        })
+        .await;
+        if !has_perm {
+            anyhow::bail!("missing CreateInvite permission");
+        }
+
         let pub_key = invite::endpoint_id_to_ed25519_public(recipient_peer_id);
 
         // Gather server info from registry.
@@ -35,10 +47,15 @@ impl<N: willow_network::Network> ClientHandle<N> {
             })
             .await;
 
-        // Get genesis author (first admin) from event_state.
+        // Use the authoritative `state.genesis_author` recorded at server
+        // creation rather than picking an arbitrary admin. Picking from
+        // `admins.iter().next()` is non-deterministic and wrong whenever
+        // additional admins have been promoted (see issue #241 / SEC-A-08).
+        // Fall back to the local peer id only for legacy states that were
+        // serialized before `genesis_author` existed.
         let my_id = self.identity.endpoint_id();
         let genesis_author = willow_actor::state::select(&self.event_state_addr, move |es| {
-            es.admins.iter().next().copied().unwrap_or(my_id)
+            es.genesis_author.unwrap_or(my_id)
         })
         .await;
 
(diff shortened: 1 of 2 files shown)


intendednull:
We are checking permissions at the client level. But if another peer has a different client (we are open source after all), this doesn't do anything right? Explain to me how this works @claude

intendednull:
Good catch — you're right that the `CreateInvite` guard in `generate_invite` is a UX guardrail, not a security boundary. Worth distinguishing the three changes in this PR because they sit at different layers.

### `generate_invite`'s `CreateInvite` check (#225 / SEC-A-02) — UX, not security

You're correct: a forked client can skip this check. The actual security boundary is the state machine: `willow-state::materialize::check_permission` requires the event author to be admin in order for `GrantPermission` to apply (see `crates/state/src/materialize.rs:103-112`). That check runs on every peer that materializes the event.

So a non-admin running a forked client can:
1. ❌ Sign and broadcast a `GrantPermission` event — but every honest peer rejects it at apply time. No state change. Recipient never gets `SendMessages`.
2. ❌ Ship an invite payload — but when the recipient redeems it, they trigger a `JoinRequest` → the inviter (still non-admin) emits `GrantPermission` → it gets rejected → recipient is in a confusing limbo where they "joined" but can't send.

The honest authorization is enforced regardless. What the new check buys us is purely **fail-fast UX**: a non-admin running an unmodified client gets a clear error immediately instead of producing an invite that silently wedges the recipient. It also matches the guard already present at `create_join_link:218-225`, so the codebase becomes consistent.

I should reframe the audit issue: **#225 is a UX/consistency fix, not a security fix.** The "high severity / security" label was overcalibrated. I'll downgrade and re-label the issue.

### `JoinRequest` signer check (#239 / SEC-A-03) — real security fix

This one **is** a security boundary, even if no client did the check. The wire packet is signed; `unpack_wire` returns the verified `signer`. Old code:

```rust
let granted_peer = peer_id;          // attacker-supplied!
// ... emits a GrantPermission with target = peer_id, signed by the inviter
```

The inviter is admin, so their `GrantPermission` event **is** accepted by every peer at apply time. Concrete attack:

1. Mallory (no relationship to the server) discovers a leaked `link_id`.
2. Mallory crafts `JoinRequest { link_id, peer_id: <anyone-Mallory-wants-to-mess-with> }` and signs it with her own key.
3. Inviter receives → `unpack_wire` returns `(JoinRequest, mallory_signer)` — but the old listener uses `peer_id` from the wire, not `signer`.
4. Inviter (admin) emits `GrantPermission { target: <anyone>, permission: SendMessages }`.
5. Every peer accepts the grant. Mallory has caused the inviter to grant permission to an arbitrary endpoint that never authenticated.

The fix uses `signer` everywhere, so Mallory can only ever grant permission to herself, and only if the link is unused/unexpired. The state-layer check doesn't catch this because the inviter genuinely holds admin — the bug is at the boundary between "what the wire claims" and "what the signature actually authenticates."

This is the real reason this PR exists; the other two are along for the ride.

### `genesis_author` (#241 / SEC-A-08) — correctness, not security

`admins.iter().next()` returns the lexicographically smallest admin (`BTreeSet`). After admins are added/removed, that may not be the original owner. The invite's `genesis_author` field is what the joiner uses as their initial trust root, so picking the wrong one steers their first sync to a non-owner admin. Not a security bypass — invites are signed hints anyway — but a correctness fix.

### Recommendation

If you want, I can:
- **Keep this PR as-is** but I'll re-label #225 and #241 to drop the `security` tag (#239 stays).
- **Or split** — drop the `CreateInvite` check entirely and keep only #239 + #241 here. I'd argue against that: fail-fast on a non-admin's invite saves a confused user, and matching `create_join_link` is a small win.

Lean toward keeping the PR, but happy to do whatever you prefer.

---

Issue #311, opened 2026-04-25
[SEC-A-09] Agent HTTP server defaults to TokenScope::Full with no CLI flag


intendednull:
**Audit finding from #300 (commit 679f9fe)**

**Severity:** medium
**Category:** auth / authorization defaults
**File:** crates/agent/src/main.rs:155
**Obvious fix:** yes

## Description
`serve_http(client, &cli.bind, Default::default(), token)` passes `TokenScope::default()` which is `TokenScope::Full` — every tool (`kick_member`, `create_server`, `delete_message`, …) is reachable from any client that holds the bearer token. There is no `--scope` flag exposed in `Cli`, so an operator deploying the HTTP transport has no way to restrict to e.g. `Messaging` or `ReadOnly` short of editing the binary.

## Impact / Threat
Operators reasonably assume "default = least privilege"; in fact default = full admin. Combined with SEC-A-01 (timing-attackable token compare) and SEC-A-08 (world-readable identity), the blast radius of any one compromise is the entire agent identity's authority on the server.

## Suggested fix
Add `--scope readonly|messaging|full|admin` (default `messaging` or `readonly`) to `Cli`, parse into `TokenScope`, pass through. Document the per-scope tool list.

## Verify
```bash
rg -n "Default::default\(\)" crates/agent/src/main.rs
```

---

**Resolved by #389** (merged 2026-04-26): adds `--scope {messaging,read,full}` clap flag, default `Messaging`, threaded into `serve_http`. Auto-close keyword chain (`closes #301 #304 #305 #311`) only fired for the first id, leaving this issue stuck open. Manually closing now.

`admin` CLI variant + regression test pinning the default were proposed in #412 but the maintainer chose to drop the increment. PR #412 closed.

Issue #301, opened 2026-04-25
[SEC-V-01] Bearer token compared with non-constant-time `==`


intendednull:
**Audit finding from #300 (commit 679f9fe)**

**Severity:** low (informational; auth was out of scope for the validation/DoS pass — see SEC-A-01 for the medium-severity version)
**Category:** security / timing oracle
**File:** crates/agent/src/server.rs:235
**Obvious fix:** yes

## Description
`bearer_auth_middleware` does `if provided == expected_token { ... }` — Rust's default `&str` equality is not constant-time, enabling a byte-wise timing oracle.

## Impact / Threat
Network-adjacent attacker can mount a timing oracle to recover the agent's MCP bearer token character by character.

## Suggested fix
Compare via `subtle::ConstantTimeEq::ct_eq(provided.as_bytes(), expected_token.as_bytes())`, or hash both sides and compare fixed-length digests.

## Verify
```bash
rg "provided == expected_token" crates/agent/src/server.rs
```

Issue #305, opened 2026-04-25
[SEC-A-02] Agent `read_resource` does not enforce `allows_resource` scope


intendednull:
**Audit finding from #300 (commit 679f9fe)**

**Severity:** low
**Category:** auth / authorization
**File:** crates/agent/src/server.rs:161
**Obvious fix:** yes

## Description
`list_resources` filters by `scope.allows_resource`, but `read_resource` skips the scope check entirely and dispatches to `resources::read_resource` directly. Today `allows_resource` always returns true so this is dormant, but as soon as any scope returns false (e.g. to hide `willow://server/join-links` from a Messaging-scoped token), a client that knows the URI can still read it because `list_resources` is the only gate.

## Impact / Threat
A future scope-tightening change is silently bypassed by clients that probe URIs directly; non-listed but readable resources are exfiltrable.

## Suggested fix
In `call_tool`-style fashion, gate `read_resource` on `self.scope.allows_resource(&request.uri)` before delegating, returning `INVALID_REQUEST` otherwise.

## Verify
```bash
rg -n "allows_resource" crates/agent/src
```

Issue #304, opened 2026-04-25
[SEC-A-01] Bearer token compared with non-constant-time equality


intendednull:
**Audit finding from #300 (commit 679f9fe)**

**Severity:** medium
**Category:** auth / timing oracle
**File:** crates/agent/src/server.rs:235
**Obvious fix:** yes

## Description
`bearer_auth_middleware` does `if provided == expected_token { ... }` — Rust's default `&str` equality is not constant-time. An attacker with timing access to the agent's HTTP endpoint can mount a byte-wise oracle to recover the token character by character.

## Impact / Threat
Network-adjacent attacker recovers MCP bearer token by timing repeated requests, gaining full agent control (send messages, manage channels, kick, etc., depending on scope).

## Suggested fix
Compare via `subtle::ConstantTimeEq` (or `constant_time_eq` crate). Compare the byte slices of equal length; first reject by length-mismatch only after a fixed-cost compare, or by hashing both and comparing fixed-length digests.

## Verify
```bash
rg "provided == expected_token" crates/agent/src/server.rs
```

PR #412, opened 2026-04-27
fix(agent): expose --scope admin + regression test for SEC-A-09

intendednull:
## Why

PR #389 closed bulk of #311 (SEC-A-09): added `--scope {messaging,read,full}`, default `Messaging`, threaded into `serve_http`. Two gaps remained.

1. Issue spec listed `readonly|messaging|full|admin`. `admin` value not exposed on CLI even though `TokenScope::Admin` exists in `scopes.rs` (semantically distinct from `Full` for audit log consumers).
2. No test pinned the regression — nothing prevents future drift back to `--scope full` default.

## What

- Add `Admin` variant to `ScopeArg`. Maps to `TokenScope::Admin`.
- Doc on `ScopeArg` now lists per-scope tool sets.
- New `mod tests` in `crates/agent/src/main.rs`:
  - `default_scope_is_messaging` — pins SEC-A-09 fix.
  - parse coverage every documented value (`messaging`/`read`/`full`/`admin`).
  - unknown-value-rejected case.
  - `ScopeArg` → `TokenScope` wiring map.

Lowest tier covering behaviour per CLAUDE.md: clap `Cli::parse_from` on the bin. No process spawn needed.

## Behaviour change

None vs current `claude/adoring-euler-3axK7`. PR #389 already flipped HTTP default `Full` → `Messaging`. This PR only adds `admin` value (additive) and locks the existing default with a test.

## Tradeoff considered

Considered flipping `TokenScope::default()` from `Full` to `Messaging` to kill the footgun globally. Rejected. Stdio transport (`serve_stdio` uses `WillowMcpServer::new` which reads the `Default`) is local-process — `Full` appropriate. Flipping `Default` would break `tests/e2e.rs` silently or force every test to `with_scope(TokenScope::Full)`. Worth a follow-up issue to remove the `Default` impl entirely and force each call site explicit.

## Test plan

- [x] `cargo fmt --check` clean
- [x] `cargo clippy -p willow-agent --all-targets -- -D warnings` clean
- [x] `cargo test -p willow-agent --bin willow-agent` 7/7 new tests pass
- [x] `cargo test -p willow-agent --lib` 31/31 pass
- [x] `cargo test -p willow-agent --test e2e` 32/32 pass

Refs #311

https://claude.ai/code/session_014KLEYzesqZwWdFQnUeXtrJ

---
_Generated by [Claude Code](https://claude.ai/code/session_014KLEYzesqZwWdFQnUeXtrJ)_

### DIFF
--- a/crates/agent/src/main.rs
+++ b/crates/agent/src/main.rs
@@ -72,7 +72,16 @@ struct Cli {
 }
 
 /// CLI-friendly enum for `--scope`. Maps to `TokenScope`.
-#[derive(Copy, Clone, Debug, ValueEnum)]
+///
+/// Per-scope tool list:
+/// - `messaging` (default): `send_message`, `send_reply`, `edit_message`,
+///   `delete_message`, `react`, `pin_message`, `unpin_message`, `send_typing`.
+///   All resources readable.
+/// - `read`: no tools. All resources readable.
+/// - `full`: every tool reachable from the bearer token, all resources.
+/// - `admin`: every tool, all resources. Semantically distinct from `full`
+///   so audit log consumers can tell intent apart.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, ValueEnum)]
 enum ScopeArg {
     /// Messaging tools only (least-privilege default).
     Messaging,
@@ -80,6 +89,8 @@ enum ScopeArg {
     Read,
     /// Full access: all tools, all resources.
     Full,
+    /// Admin access: all tools, all resources (audit-distinct from `full`).
+    Admin,
 }
 
 impl From<ScopeArg> for TokenScope {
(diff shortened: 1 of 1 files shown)


intendednull:
Is this still needed?

intendednull:
good catch. #389 already shipped `--scope {messaging,read,full}` + Messaging default + thread into `serve_http`. issue auto-close keyword chain didn't fire (`closes #301 #304 #305 #311` — GH only closed first id), so #311 stuck open.

what #412 actually adds:
1. `admin` CLI variant — issue spec listed `readonly|messaging|full|admin`, `admin` was missing. `TokenScope::Admin` already exists in `scopes.rs` but unreachable from CLI.
2. regression test pinning `default_scope_is_messaging` so default can't drift back to `Full`.

3 options:
- **keep both** — fills spec gap + locks default. ~95 lines, all in `main.rs`.
- **keep just test** — drop `admin` variant, just pin default.
- **close + manual-close #311** — declare #389 sufficient, ship nothing here.

your call. defaulting to close if no preference.

---

intendednull:
Clean up please

Issue #448, opened 2026-04-27
Search index: thread active-letter id into IndexableMessage.letter_id


intendednull:
Follow-up to #355.

`crates/web/src/app.rs` builds `IndexableMessage` from the messages signal and currently hard-codes `letter_id: None` because there is no active-letter signal in `crates/web/src/state.rs` to read from. Letters/DMs is specced in `docs/specs/2026-04-19-ui-design/letters-dms.md` but the feature has not landed:

- No `letter_id` / `current_letter` signal in `AppState`.
- `command_palette` + `tab_bar` reference letters as a feature-flagged future surface.
- `SearchScope::ThisLetter(id)` and `SearchScope::AllLetters` exist in the executor and rely on `Posting.letter_id`, but no code path produces a non-`None` `letter_id` today.

When letters-dms.md ships:

1. Add `current_letter: ReadSignal<Option<String>>` (or equivalent) to `AppState`.
2. Subscribe to it in the search-index Effect in `crates/web/src/app.rs` alongside `current_channel` / `active_server`.
3. Set `letter_id: current_letter.get()` (or per-message lookup if the messages signal mixes channels and letters).
4. Add a `client`-tier test confirming `SearchScope::ThisLetter` returns the expected hits.

Unblocks the documented `in:#letter` / scope-flip operator behaviour. Severity: medium (UX gap, not security).

intendednull:
Duplicate of #441 — same problem (`letter_id: None` hard-coded in indexer effect, blocked on letters-dms feature). Consolidate work under #441.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #355, opened 2026-04-25
[GEN-12] Search index drops `letter_id`, `has_image`, `has_file` — operators silently underreport


intendednull:
**Audit finding from #300 (commit 679f9fe)**

**Severity:** medium
**Category:** bug
**File:** crates/web/src/app.rs:382
**Obvious fix:** no

## Description
When constructing `IndexableMessage` in app.rs the indexer hard-codes `letter_id: None`, `has_image: false`, `has_file: false`. The search executor's `has:image` / `has:file` / `in:#letter` operators therefore never match anything, despite being part of the public `SearchScope` / parser API and surfaced in spec text.

## Impact / Threat
User-facing — documented search operators silently return zero results; users assume their files / images aren't searchable.

## Suggested fix
Derive `has_image` / `has_file` from `m.attachments` (or whatever DisplayMessage exposes) and plumb the letter id from the active letter context.

## Verify
```bash
rg "letter_id: None,|has_image: false,|has_file: false," crates/web/src/app.rs
```

intendednull:
audit sweep 2026-04-26 @ 2f26d91: search component absent on `main`. finding still valid against feature branch; revisit when merged. leaving open.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #449, opened 2026-04-27
fix(web): plumb has_image/has_file into search index

intendednull:
## Summary

- Indexer at `crates/web/src/app.rs:382` hard-coded `has_image: false` / `has_file: false` / `letter_id: None`, so the documented `has:image` / `has:file` / `in:#letter` operators silently returned zero hits.
- New pure helper `willow_client::derive_has_image_file(body)` derives the booleans from the body string: inline `[file:name:b64]` payloads (the format produced by `ClientHandle::share_file_inline`) split on filename extension, and image URLs in plain text fire `has:image` to match the renderer's auto-embed behaviour.
- Wired the helper into the indexer Effect in `app.rs`. `letter_id` stays `None` for now — there is no active-letter signal in web state because letters-dms.md hasn't landed. Tracked in #448.

## Tradeoffs

- Helper lives in `willow-client` (not `willow-web`) because the inline-file body format is owned by `ClientHandle::share_file_inline` and a future native consumer (agent / CLI) needs the same booleans. Runner-up: parse-from-`Content::File` directly — rejected because `DisplayMessage` is built from `ChatMessage` which only carries `body: String`; the structured `Content` enum is collapsed at projection time, so body is the only signal available without a wider refactor.
- Image URL detection mirrors the renderer's extension-list rule rather than full URL parsing — keeps the indexer cheap and matches what users visually see embedded.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo check -p willow-web --target wasm32-unknown-unknown`
- [x] `cargo test -p willow-client --lib search::` (10 new tests in `derive::tests` + 1 regression test in `execute_tests::derived_attachment_flags_drive_has_image_and_has_file_operators`)
- [x] `cargo test -p willow-web --lib`

Refs #355
Follow-up: #448

https://claude.ai/code/session_01HnwK1dP6rnrsbLxtg92zfF

---
_Generated by [Claude Code](https://claude.ai/code/session_01HnwK1dP6rnrsbLxtg92zfF)_

### DIFF
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -79,8 +79,8 @@ pub use nickname::{MemNicknameStore, NicknameStore, NicknameStoreHandle, NICKNAM
 pub use ops::{pack_wire, unpack_wire, VoiceSignalPayload, WireMessage};
 pub use queue::{ArrivedSummary, QueueSummary, RelayStatus};
 pub use search::{
-    IndexableMessage, RecentQuery, SearchIndex, SearchIndexBuildStatus, SearchIndexConfig,
-    SearchIndexHandle, SearchQuery, SearchResult, SearchScope,
+    derive_has_image_file, IndexableMessage, RecentQuery, SearchIndex, SearchIndexBuildStatus,
+    SearchIndexConfig, SearchIndexHandle, SearchQuery, SearchResult, SearchScope,
 };
 pub use trust::{
     ComparePreview, InMemoryTrustStore, PeerTrust, TrustStore, TrustStoreHandle, UnverifiedReason,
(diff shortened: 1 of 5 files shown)


intendednull:
Fix conflicts please

intendednull:
Superseded by #442. main now have `IndexableMessage::from_display_message` with same has_image/has_file derive logic. Closing — fix already shipped.

Test coverage on main covers same surface (search/tests.rs `from_display_message_tests` mod). Letter_id plumbing follow-up still tracked at #448.

---

Issue #493, opened 2026-04-29
general-audit lessons: 2026-04-29


intendednull:
Lessons from `/general-audit` run @ `0de7631` (#492). Built on lessons #426 + #438 + #477.

## What worked (from #477 suggestions, now validated)

- **Lesson #1 applied: `let _ = .<method>().await` in standard sweep grep set.** Verified F2 from #474 stayed fixed — zero hits across `crates/web/src/components/`, `handlers.rs`, `app.rs`. Sweep correctly found nothing today (good — means the fix held), but the sweep itself was cheap and ran in seconds.
- **Lesson #2 applied: explicit sibling-of-closed pass.** This was the lesson that produced both findings today. Both F1 + F2 are titled `ci:` in commit subject but never edited `.github/workflows/`. Standard sweep set wouldn't have caught either — they're CI-wiring gaps, not code patterns. Without this lesson, this audit would have had 0 findings.
- **Lesson #3 applied: codified `mcp__github__sub_issue_write` linking.** Wired F1 + F2 as sub-issues of master immediately after creation. Surfaces them in master's UI panel without manual cross-ref.
- **Lesson #5 applied: bumped threshold from #438's "<50 files" to "<100 files / <2000 LOC".** Today's diff was 66 files / +12733 LOC — well over LOC ceiling but orchestrator-direct still scaled because finding density was low (test-hooks scaffolding + test landings, both well-gated and following spec). Real signal is finding density per file, not file count alone. Threshold rule held.
- **Lesson #6 applied: standard sweep grep set in skill body.** Ran it as a pre-commit checklist on first pass — `unsafe`, `dbg!`/`todo!`/`unimplemented!`/`FIXME`/`HACK`, locks-without-lock-ok, `js_sys::eval`/`innerHTML`, `let _ = .await`, anyhow in pure libs, docker pin, `cargo audit`. All clean. Confirmed strong areas in minutes vs reconstructing from memory.
- **Pre-fetched existing-issue lists for dedup** (lesson #4 from #438). 60 unique titles across audit + tech-debt + security labels. Inline dedup caught 9+ already-tracked items immediately (WS-1, TD-14, SEC-V-02/05/06/07, TD-02/03 structural, all cargo-audit warnings).

## What didn't

[...]

Issue #477, opened 2026-04-28
general-audit lessons: 2026-04-28


intendednull:
Lessons from `/general-audit` run @ `958e1ec` (#474). Built on lessons #426 + #438.

## What worked (from #438 suggestions, now validated again)

- **Lesson #2 applied: orchestrator-only sweep at the threshold.** 52-file diff = right at #438's "<50 file" suggested cutoff. Direct sweep still scaled — found 2 real findings (F1 Docker trunk pin, F2 component handler swallow) without agent context blow-up. Shape held.
- **Lesson #4 applied: pre-fetched existing-issue lists.** 69 unique titles (audit + tech-debt + security labels combined). Inline dedup caught 9 dup candidates immediately (AUD-1/AUD-2 already fixed via #445/#450, WS-1 #425, TD-14 #332, SEC-V-06 #235, SEC-V-07 #236, DEP-02 #313, all cargo-audit warnings).
- **Lesson #5 applied: orchestrator ran cargo-audit directly.** Clean vs CI ignore list. Zero agent needed. (Crates.io 403 on yank-check was harmless — vuln scan completed exit 0.)
- **F2 root-cause matched #350's class.** Sibling issues across closed-fix-scope mismatches keep surfacing — last run's AUD-2 was that pattern (sec-authperm scope-gate forgotten in `read_resource`). This run: handlers.rs got `warn_and_toast` but components/ was missed. Pattern: closing an issue against one file leaves siblings in adjacent files.

## What didn't

- **F2 only surfaced because I ran a final "swallowed errors in new code" sweep.** The standard sweep set (unsafe/dbg/eval/anyhow/lock-ok/Vec caps) wouldn't have caught it; `let _ = h.<method>().await` is a UX/observability pattern, not a security one. Add to standard sweep grep set.
- **No sub-issue linking step in the skill** — I had to call `mcp__github__sub_issue_write` manually after creating issues. Worth codifying.
- **No verification of "did the fix in PR X actually resolve the open issue" step.** Audit skill currently treats closed issues as fixed; but PRs sometimes close issues with partial fixes (e.g. #319 closed with deploy.yml-only fix, leaving Docker pipeline scope unaddressed → F1 today). Sibling-of-closed pattern needs explicit hunting.
[...]

Issue #438, opened 2026-04-27
general-audit lessons: 2026-04-27 (run 2)


intendednull:
Lessons from `/general-audit` run @ `00aa515` (#437). Second run on same day as #413 (yesterday); built on lessons from #426.

## What worked (from #426 suggestions, now validated)

- **Lesson #3 applied: skipped 8-agent fan-out entirely.** Single orchestrator-direct sweep. Result: 2 real findings filed in minutes. No agent timeouts (because no agents). Trade-off acceptable for small diff.
- **Lesson #4 applied: pre-fetched existing-issue lists** (`/tmp/audit-issues.txt` 17 entries, `/tmp/all-quality-issues.txt` 89 entries). Dedup happened inline without per-grep round trips. Caught 5 dup candidates immediately (#425, #332, #235, #311, deps cluster).
- **Lesson #5 applied: orchestrator ran cargo-audit directly.** All 5 warnings already in CI ignore list w/ tracked issue numbers. Zero agent needed.
- **Direct grep sweep covered the same surface as last run's 8 agents** for a tiny diff. Patterns swept: unsafe, dbg!/eprintln!, locks-without-lock-ok, panic! in prod, eval/innerHTML, TODO/FIXME, anyhow in pure libs, Vec caps, .ok() swallow.

## What didn't

- **Diff was so small (6 commits, 4 files) that the audit was mostly redundant w/ #413.** Same-day re-runs produce diminishing returns. AUD-2 (TokenScope::ReadOnly leaks join_links) was missed by #413's 8-agent run despite being unchanged code — proving the value of *some* fresh sweep, but the 1-day cadence is overkill.
- **AUD-2 should have been caught by yesterday's `sec-authperm` agent** but that agent was one of the 7 that timed out. Lesson: when agents timeout, the gaps are real and need backfill — not just absorbed into "lessons learned."
- **Worktrees skipped** since no agents → setup time saved but no isolation for parallel runs if I'd needed them.

## Suggested edits to `.claude/skills/general-audit/SKILL.md`

1. **Add a "skip-window" rule.** "If HEAD is within N commits AND M hours of last audit master issue, skip — diff-only review (current PR rules apply) instead of full re-audit. Suggest N=20, M=24h." Today's run produced 2 findings on a 4-file diff; full sweep mostly re-walked yesterday's ground.

[...]

Issue #426, opened 2026-04-27
general-audit lessons: 2026-04-27


intendednull:
Lessons from `/general-audit` run @ 401e2fc (#413).

## What worked

- **Worktree-per-agent setup** + pre-installed `cargo-audit` + report scaffolding. Infrastructure layer was solid.
- **Direct main-context audit** w/ rg + targeted Read filled gaps after agents stalled. Yielded WS-1 finding + verified state-crate purity, wire deser caps, lock-ok comment compliance.
- **Second-pass grep verification** caught 3 invalid findings before file (F4 ChannelRevive test exists, F11 dup #340, F13 dup #232). Saved 3 noise issues.
- **Existing-issue dedup** before file. 5 prior issues caught (#223 #246 #247 #316 #317 #318 deps; #340 #341 #343 #232 test).
- **testcov agent** survived w/ 13 detailed findings; was lone successful subagent.

## What didn't

**Catastrophic failure mode: 7/8 subagents stream-idle-timeout before writing findings.**

- Agents exhausted 30–60+ tool calls on exploration, then hit upstream API stream-idle timeout at finding-write time.
- Resulting reports = scaffold header only, zero findings written.
- Pattern: agents explored breadth-first across whole tree, accumulating context, then attempted one large batched write at the end. Write never happened.

The "scaffold before 2nd tool call" instruction worked (scaffolds exist) but the "append each finding fully formed before next" instruction failed — agents never produced even one fully-formed finding.

## Root cause hypothesis

Agents are coached on incrementality but not enforced. Long context window + many tool calls + final-write batching = vulnerable to upstream timeout at exactly the wrong moment.

## Suggested edits to `.claude/skills/general-audit/SKILL.md`

1. **Hard cap on tool-call budget per agent before first finding written.** Add: "If the agent makes >5 tool calls without appending one finding to the report file, STOP exploring — write the strongest one finding seen so far, then continue. No exception."

2. **Reduce agent concern surface.** 6 min × broad ("input validation across whole tree") was wrong split. Split per-crate or per-narrow-pattern. E.g. split sec-input into "deser caps in transport+common" + "panics in state machine" + "rate-limits in relay." Each agent then has narrow grep universe.

[...]

PR #494, opened 2026-04-29
docs(skill): fold audit lessons #426 #438 #477 #493 into general-audit

intendednull:
Promote sibling-of-closed pass to first (100% of recent findings), default
orchestrator-direct sweep <100 files / <2000 LOC, drop 8-agent fan-out,
add concrete sweep grep set + scope-vs-claim + API-without-caller patterns,
codify search_issues over list_issues for >25-issue labels and
sub_issue_write linking, and add a self-improvement step that opens a PR
folding each lessons-issue's suggested edits back into this skill file.

https://claude.ai/code/session_016K2qwUf2o9hgge41LXYUXK

### DIFF
--- a/.claude/skills/general-audit/SKILL.md
+++ b/.claude/skills/general-audit/SKILL.md
@@ -73,21 +73,9 @@ git diff <pr-base>..<pr-head> -- '*.rs' | rg "^\+\s*pub (async )?fn (\w+)" -or '
 # Then for each fn name: rg "\.<fn_name>\(" crates --glob '!**/tests*'
 ```
 
-### Pass 2: Pre-fetch existing-issue lists
+### Pass 2: Standard sweep grep set
 
-**Critical:** use `mcp__github__search_issues` (not `list_issues`) when a label has ≥25 open issues — `list_issues` exceeds the 78k-char tool-result cap and the workaround (save-to-file + jq) is slow.
-
-```text
-mcp__github__search_issues "is:open repo:intendednull/willow label:audit"
-mcp__github__search_issues "is:open repo:intendednull/willow label:security"
-mcp__github__search_issues "is:open repo:intendednull/willow label:tech-debt"
-```
-
-Hold the union of titles as the inline dedup index for the rest of the audit. Pre-fetch BEFORE any agent spawns, BEFORE any greps run.
-
-### Pass 3: Standard sweep grep set
-
-Run as a checklist; surface anything new vs the dedup index:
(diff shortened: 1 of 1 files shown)


.claude/skills/general-audit/SKILL.md
```diff
@@ -6,60 +6,182 @@ user-invocable: true
...
-Spawn more if area needs depth.
+**Default to orchestrator-only sweep** when diff < 100 files AND < 2000 LOC since last audit master commit.
+
+History across runs #437, #474, #492: orchestrator-direct produced all findings; agent fan-out (#413) had 7/8 stream-idle timeouts and 1 successful agent. Default = no agents.
+
+Spawn agents **only when**:
+- Diff > 100 files OR > 2000 LOC, **AND**
+- Specific concern needs deeper grep than single context can hold.
+
+When agents fan out: split **per-crate** (one crate, all concerns). Never per-concern (one concern, whole tree) — per-concern blew context window in #413.
```
intendednull: I'm not so sure about this. I think I still want sub-agent reviews. We don't want to assume there are fewer issues since the last run. We want to make a full sweep every time with all the resources every time.

.claude/skills/general-audit/SKILL.md
```diff
@@ -6,60 +6,182 @@ user-invocable: true
...
+grep -rn "FROM [^@]*$" docker/   # any unpinned base image?
+```
+
+### Pass 3: cargo-audit ignore-list drift
+
+After `cargo audit` completes, diff the CI `--ignore` list against the current advisory DB. Stale RUSTSEC IDs that no longer match should be surfaced as low-priority cleanup ("N stale RUSTSEC IDs in ci.yml — can be pruned"). Cosmetic but accumulates.
+
+### Pass 4 (only if agents spawned): timeout backfill
+
+If any agent timed out without writing findings, the orchestrator MUST manually sweep that concern before declaring the audit complete. Gaps from timed-out agents compound across runs (AUD-2 sat undetected for one full audit cycle in #413 → #437 because nobody backfilled the `sec-authperm` agent's concern).
```
intendednull: A lot of timeouts happen when we are trying to write big chunks of data. We should include instructions to write in small chunks so we don't time out as often

intendednull (commented):
Take this time to review and reorganize and approve if you see anything that can be improved

intendednull:
Also let's not mention previous audit-like specific links or anything like that. We don't want to confuse future agents

